PR #15, opened 2023-05-25
Update Dockerfile

galindus:
Use base image with compatible torch cuda version (11.7)


justinmerrell:
@zhl146 From my tests with the runtime image, I can confirm that 11.7 is more reliable, with 11.7 being on their stable list. 

https://pytorch.org/blog/deprecation-cuda-python-support/

zhl146:
Will update this soon - waiting because we have a ton of machines stuck on 11.6 :(

justinmerrell:
@zhl146 I believe the changes in this PR are backwards compatible

justinmerrell:
Containers updated

PR #21, opened 2023-07-11
Added H20gpt template

kodxana:


justinmerrell:
May you modify it to work with the same start file as our other containers? 
Also, what ports do we need to open in the template?

PR #23, opened 2023-07-22
Update nginx.conf

slmagus:
allows /upload of at least 1GB. This could probably be increased.

File uploads were failing with a HTTP 413 even when we bypass the cloudflare reverse proxy this setting enables files upload to all routes on the webUI. If a file larger then the max body size is uploaded, the stream is cut off when it reach the max body size

Tested on slmagus/stable-diff:dev which required some hacks to bypass the lack of a proxy image. This merge assumes the updates will be included when --from=proxy is rebuilt.

### DIFF
--- a/container-template/proxy/nginx.conf
+++ b/container-template/proxy/nginx.conf
@@ -1,7 +1,7 @@
 events { worker_connections 2048; }
 
 http {
-
+    client_max_body_size 1024M;
     # invokeai
     server {
         listen 9091;
@@ -27,7 +27,7 @@ http {
     # Fast Stable Diffusion + web UI + Comfy UI
     server {
         listen 3001;
-	client_max_body_size 1024M;
+		
         location /ws {
             proxy_http_version 1.1;
             proxy_set_header Accept-Encoding gzip;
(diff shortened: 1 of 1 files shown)


container-template/proxy/nginx.conf
```diff
@@ -27,7 +27,7 @@ http {
     # Fast Stable Diffusion + web UI + Comfy UI
     server {
         listen 3001;
-
+	client_max_body_size 1024M;
```
ashleykleynhans, line 30: Why do you want to set this to 1GB and not just 20MB which should be more than sufficient?
It would also be better to place this in the `http` section than `server ` section so all the other containers can also benefit from it.

slmagus: I want to be able to upload larger video files. to use with vid2vid type extensions.

I moved the setting to the server section.

I tested with and without going through the cloud flare proxy with this setting.
  - with the proxy, files up to 100MB seem to be allow, I assume that is due to the cloudflare setting.
  - without the proxy, and by directly exposing the tcp port during pod setup, I am able to upload files up to 1024MB

justinmerrell (approved):

PR #45, opened 2023-11-13
Change behaviour on how ssh keys are added to authorized_keys

Fluder-Paradyne:
This PR updates the startup script to overwrite the `authorized_keys` file with the provided `PUBLIC_KEY` environment variable, rather than appending to it. This change ensures that only the specified key will be authorized and does not accumulate keys from successive reboots.

### DIFF
--- a/container-template/start.sh
+++ b/container-template/start.sh
@@ -26,7 +26,7 @@ setup_ssh() {
     if [[ $PUBLIC_KEY ]]; then
         echo "Setting up SSH..."
         mkdir -p ~/.ssh
-        echo "$PUBLIC_KEY" >> ~/.ssh/authorized_keys
+        echo "$PUBLIC_KEY" > ~/.ssh/authorized_keys
         chmod 700 -R ~/.ssh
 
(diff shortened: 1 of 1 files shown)


Fluder-Paradyne:
image

After restarting the pod

image

Fluder-Paradyne:
@justinmerrell
Hi can I get some comments on this.

justinmerrell:
Restarting the pod should reset the keyfile.... I think, also pretty late but can verify this tomorrow. 

As far as over-writing, users do pre-populate keyfiles when creating docker image and then might decide to add more later via their settings. 

Let me know if I overlooked the problem that this PR is seeking to resolve.

Fluder-Paradyne:
sorry for late reply,
caught up in work.

I was having ssh issues, thought the duplication of keys was the cause, but it seems to not the cause. ( it was my ssh-agent )
Keys are getting duplicated on restarts but not causing issues.

justinmerrell:
Sounds like this PR is not required.

PR #59, opened 2024-04-14
Add runpod/cuda image (12.4, 12.3, 12.2, 12.1, 11.8)

alpayariyak:
We don't have any cuda 12.2 and 12.3 images that users can SSH into and use the Jupyter notebook functionality with, while the demand for those cuda versions keeps rising

### DIFF
--- a/official-templates/cuda/Dockerfile
+++ b/official-templates/cuda/Dockerfile
@@ -1,0 +1,66 @@
+ARG BASE_IMAGE
+FROM ${BASE_IMAGE}
+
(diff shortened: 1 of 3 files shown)


yocontra:
Can you align https://github.com/runpod/containers/blob/main/official-templates/base/docker-bake.hcl with this so they have the same versions?

alpayariyak:
@yocontra thank you for the suggestion, added those versions in #63.

max4c:
Closing — stale for 2+ years with no activity.

PR #60, opened 2024-04-16
FIX fast stable diffusion pod template TMALLOC ( memory problem fix )

nerdylive123:
like on https://github.com/AUTOMATIC1111/stable-diffusion-webui/discussions/6722#discussioncomment-5609254.
This is a fix for memory problem when user switched or loaded models/lora multiple time, then the SD will crash after.

pls do tests on this fix, i don't use the minimal version, i usually use the normal version of libtcmalloc.

### DIFF
--- a/official-templates/fast-stable-diffusion/Dockerfile
+++ b/official-templates/fast-stable-diffusion/Dockerfile
@@ -8,6 +8,7 @@ WORKDIR /
 ENV SHELL=/bin/bash \
     PYTHONUNBUFFERED=1 \
     DEBIAN_FRONTEND=noninteractive
+    LD_PRELOAD=/usr/lib/x86_64-linux-gnu/libtcmalloc_minimal.so.4
 
 # Set up system
 RUN apt-get update --yes && \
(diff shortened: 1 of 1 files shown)


official-templates/fast-stable-diffusion/Dockerfile
```diff
@@ -8,6 +8,7 @@ WORKDIR /
 ENV SHELL=/bin/bash \
     PYTHONUNBUFFERED=1 \
     DEBIAN_FRONTEND=noninteractive
```
ashleykleynhans, line 10: You are missing a slash here.

ashleykleynhans (changes requested):

ashleykleynhans:
> like on [AUTOMATIC1111/stable-diffusion-webui#6722 (comment)](https://github.com/AUTOMATIC1111/stable-diffusion-webui/discussions/6722#discussioncomment-5609254). This is a fix for memory problem when user switched or loaded models/lora multiple time, then the SD will crash after.
> 
> pls do tests on this fix, i don't use the minimal version, i usually use the normal version of libtcmalloc.

Minimal version works fine, but I wonder why you're installing the deb packages from TheLastBen's Github repo rather than simply installing the Ubuntu apt packages.

official-templates/fast-stable-diffusion/Dockerfile
nerdylive123: Thank you for the suggestion, i didn't see that

nerdylive123:
Well yeah, idk i think it would be great to use the deps from that repo

ashleykleynhans:
> Well yeah, idk i think it would be great to use the deps from that repo

I would agree with you if they were related to the repo itself, but its pretty bad practice to use debian packages that are version locked and very likely outdated.  Its best practice to install the latest packages from the Operating System to ensure that you get the latest security patches included.

ercanozer:
@nerdylive123 @ashleykleynhans When can we get this merged?

max4c:
Closing — stale for 2+ years with no activity.

PR #68, opened 2024-08-19
add template for pytorch 2.4

Yhlong00:
## Description: 

- Added PyTorch 2.4 support for both Nvidia and AMD GPUs.
- Sorted the lines for better organization.

## How to Test?

- Built the Docker image and pushed it to Docker Hub.
- Tested the image by running it on RunPod as a pod.
- Execute the script in the Jupyter notebook to validate the functionality of PyTorch, TorchVision, and TorchAudio

### DIFF
--- a/official-templates/pytorch/README.md
+++ b/official-templates/pytorch/README.md
@@ -1,7 +1,14 @@
-## Build Options
+## Build Instructions
 
-To build with default options, run `docker buildx bake`, to build a specific target, run `docker buildx bake <target>`.
+- To build with the default options, simply run `docker buildx bake`.
+- To build a specific target, use `docker buildx bake <target>`.
+- To specify the platform, use `docker buildx bake <target> --set <target>.platform=linux/amd64`.
 
-## Ports
+Example:
+```bash
+docker buildx bake 240-py311-cuda1240-devel-ubuntu2204 --set 240-py311-cuda1240-devel-ubuntu2204.platform=linux/amd64
+```
+
+## Exposed Ports
 
 - 22/tcp (SSH)
--- a/official-templates/pytorch/docker-bake.hcl
+++ b/official-templates/pytorch/docker-bake.hcl
@@ -17,7 +17,7 @@ group "default" {
         "220-py310-cuda1211-devel-ubuntu2204",
         "221-py310-cuda1211-devel-ubuntu2204",
         # CUDA 12.4.0
-        "240-py311-cuda1240-devel-ubuntu2204",
+        "240-py311-cuda1241-devel-ubuntu2204",
 
         ### ROCM ###
         # ROCM 5.6
@@ -59,7 +59,7 @@ group "cuda" {
         "211-py310-cuda1211-devel-ubuntu2204",
         "220-py310-cuda1211-devel-ubuntu2204",
         "221-py310-cuda1211-devel-ubuntu2204",
-        "240-py311-cuda1240-devel-ubuntu2204",
+        "240-py311-cuda1241-devel-ubuntu2204",
     ]
 }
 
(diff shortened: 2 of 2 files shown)


TimPietruskyRunPod (approved):
The changes LGTM, I will test both images on an actual machine next.

TimPietruskyRunPod:
## Testing

Tested the following GPUs successfully:

* RTX 3090
* RTX 4090

For the following GPU, I saw a warning:
* MI300x

```bash
/opt/conda/envs/py_3.10/lib/python3.10/site-packages/torchvision-0.16.1+fdea156-py3.10-linux-x86_64.egg/torchvision/io/image.py:13: UserWarning: Failed to load image Python extension: '/opt/conda/envs/py_3.10/lib/python3.10/site-packages/torchvision-0.16.1+fdea156-py3.10-linux-x86_64.egg/torchvision/image.so: undefined symbol: _ZN3c1017RegisterOperatorsD1Ev'If you don't plan on using image functionality from `torchvision.io`, you can ignore this warning. Otherwise, there might be something wrong with your environment. Did you have `libjpeg` or `libpng` installed before building `torchvision` from source?
```


## Test code

The test was done by using the following code:

```python
import torch
import torchvision
import torchaudio

def test_pytorch():
    print("Testing PyTorch...")
    # Check if CUDA is available
    cuda_available = torch.cuda.is_available()
    print(f"CUDA available: {cuda_available}")

    # Create a random tensor and move it to GPU if available
    device = torch.device("cuda" if cuda_available else "cpu")
    x = torch.rand(3, 3).to(device)
    print(f"Random Tensor:\n{x}")
    print(f"Tensor Device: {x.device}")
    
    # Perform a simple matrix multiplication
    y = torch.mm(x, x)
    print(f"Matrix multiplication result:\n{y}\n")
    
    print("PyTorch test passed!\n")

def test_torchvision():
    print("Testing TorchVision...")
    # Load a pretrained model using the updated `weights` parameter
    weights = torchvision.models.ResNet18_Weights.DEFAULT
    model = torchvision.models.resnet18(weights=weights)
    model.eval()  # Set the model to evaluation mode
    
    # Create a dummy input and pass it through the model
    dummy_input = torch.rand(1, 3, 224, 224)
    with torch.no_grad():
        output = model(dummy_input)
    
    print(f"Output from ResNet18:\n{output}\n")
    
    print("TorchVision test passed!\n")

def test_torchaudio():
    print("Testing TorchAudio...")
    # Generate a test signal
    sample_rate = 16000
    freq = 440.0
    duration = 1.0  # seconds
    t = torch.arange(0, int(sample_rate * duration)) / sample_rate
    waveform = 0.5 * torch.sin(2 * torch.pi * freq * t)

    # Compute the spectrogram
    spectrogram = torchaudio.transforms.Spectrogram()(waveform)
    print(f"Spectrogram shape: {spectrogram.shape}\n")
    
    print("TorchAudio test passed!\n")

if __name__ == "__main__":
    test_pytorch()
    test_torchvision()
    test_torchaudio()
```

maglietti:
Should `libjpeg` and `libpng` be pre-installed in the container so the user doesn't have to worry about installing them? Are the installation instructions included in the README for the template?

TimPietruskyRunPod:
I did not receive this warning for the NVIDIA image, so I think there is something missing in the AMD image. From my point of view, we should have these installed, so that users can just use everything without the need to install the basic libs themselves. 

There is no README for this template at all that users can see on RunPod.

Yhlong00:
`libjpeg` and `libpng` are primarily used for manipulating JPEG and PNG images, including tasks like compression, decompression, resizing, and cropping. If you’re working on computer vision projects, you’ll likely need these libraries. However, it’s surprising that the official ROCm AMD image doesn’t include them by default.

Installing these packages in our Dockerfile would add them to **all** our PyTorch templates, which could introduce additional risks and require thorough testing. Given the current situation, I would prefer not to include them at this time. Additionally, since fewer people are using AMD, those who do need these libraries should be able to install them without much difficulty.

And yes, we don't have README file for a lot of official template, we can add them later.

PR #92, opened 2025-05-25
Fix: Rename extra_model_paths.yml to .yaml for ComfyUI compatibility

TheMostlyGreat:
Currently, the comfy UI template isn't picking up the workspace/comfyui/models folder because the extra_model_path file has a .yml extension instead of a .yaml extension which is what comfy seems to expect. So I  renamed extra_model_paths.yml to .yaml extension to match ComfyUI's expected file extension format. And updated the docker file and the readme.


max4c:
Closing — stale for 11 months with no activity and has merge conflicts.

PR #99, opened 2025-08-19
fix: export all env variables to login shell for ssh

### LINKED ISSUE #70: Custom environment variables not export in true ssh
After creating a pod and specifying some environment variables. I can see how these are accessible from the Basic SSH and the JupyterLab Terminal. However, when doing "True SSH" these environment variables are nowhere to be seen

k-nox:
- fixes CE-225 where direct ssh connections don't have instant cluster variables
- also closes #70 so user defined custom environment variables are present in ssh sessions


DeJayDev (approved):
thank you!

BrownianNotion:
Also still getting this issue

PR #100, opened 2025-09-16
feat: containers v1

DeJayDev:
Containers v1 :)

The first major release and deployment of the new containers repo introduced with the v0.7.0 lineup in #94.

Primarily, v1:

* Fixes a longstanding issue with exporting environment variables multiple times, wiping changes to the `.bashrc` file.
* Drops building Ubuntu 20.04 based images
* Simplifies the tag offering (it was a good idea to have multiplies but forgot Docker Hub isn't normal)
* Adds slurm-wlm to base image
* Drops building images for pyTorch versions below 2.5.1, adds 2.8.0
* Drops building images for CUDA versions below 12.6.3, add 13.0.0


max4c:
Hey! Before we merge to main, I have a few questions about potential breaking changes and user communication:

Breaking Changes I noticed:
- Default Python version changed from 3.10 → 3.12 - Will this break existing user workflows/scripts that rely on python being 3.10?

- Removed post_start.sh which had: SSH inactivity monitoring (auto-removes pods after no SSH connections)
Filebrowser auto-start on port 4040
Was this removal intentional? Do we need to communicate this to users or provide an alternative?

- Dropped Ubuntu 20.04 support and older CUDA versions (12.4.1, 12.5.1, 12.6.3) - Are there users on these versions we need to notify?

- Version bump 0.7.0 → 1.0.2 - Appropriate for the changes, but do we have release notes prepared?



Aside from that, the code looks good - CI/CD improvements are clean, bug fixes are solid, and the Docker configs validate correctly. Just want to make sure we're not surprising any users!

max4c (approved):
Please ping me if any PR goes unchecked for more than 48 hours! We want to merge fixes/improvements efficiently!

PR #103, opened 2025-11-07
Vllm ray template

runpod-ebooks:
A cluster template for a VLLM + Ray workflow for distributed inference. POC for instant clusters

Requires the following environment variables to be specified:
HF_MODEL :: the model to be served, downloaded from huggingface, specified as [username]/[modelname]

Optionally, you can also set a token environment variable if you need models that are gated
HF_TOKEN :: your access token from huggingface


murat-runpod (changes requested):

max4c:
Closing — stale for 5 months with unresolved change requests.

PR #105, opened 2025-11-14
Reintroduce post_start.sh

runpod-ebooks:
Adding additional setup steps to images is harder without being able to reference environment variables in them, so I think it's reasonable to put a script we can hook to do stuff after that back in this flow


max4c:
Closing — stale for 5 months and the build would fail (post_start.sh is missing from container-template/). The hook point already exists in start.sh — reopen with the actual script file if still needed.

Issue #101, opened 2025-10-08
Pytorch 2.8.0 image issues


nFunctor:
Hello,

The relatively new `runpod/pytorch:1.0.1-cu1281-torch280-ubuntu2404` image seems to have various issues not present with other images.

Whenever I `ssh` (via the basic link, no TCP), I get a bunch of errors like `bash: export: `ssh-ed25519': not a valid identifier` that are followed by the public keys of people from my orgnaisation (the connection works though).

Downloading models via VLLM seems impossible as well: I ended up using Huggingface CLI to download the weights manually (something that is normally done automatically via `vllm serve`). 

The image has been tried with multiple GPUs (A40, RTX5090) and presents the same behaviour (not found when switching to eg Pytorch 2.4.0 image). Overall it is possible to work with it but I'd be happy to see those issues addressed. Thanks!

max4c:
Closing — root cause identified in #114 (pip silently installed wrong PyTorch version). Fix in PR #115.

Issue #98, opened 2025-08-10
Error using `pytorch:2.8.0-py3.11-cuda12.8.1-cudnn-devel-ubuntu22.04` Pod Template


Ompragash:
System Log:
```
8/10/2025, 10:09:09 PM
start container for runpod/pytorch:2.8.0-py3.11-cuda12.8.1-cudnn-devel-ubuntu22.04: begin
error starting container: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error running hook #0: error running hook: exit status 1, stdout: , stderr: Auto-detected mode as 'legacy'
nvidia-container-cli: requirement error: unsatisfied condition: cuda>=12.8, please update your driver to a newer version, or use an earlier cuda container: unknown
```

max4c:
Closing — root cause identified in #114 (pip silently installed wrong PyTorch version). Fix in PR #115.

PR #115, opened 2026-04-02
Add PyTorch version verification to prevent silent pip fallback

### LINKED ISSUE #114: PyTorch 2.8.0 cu128 template installs 2.4.1 — missing cu128 wheels
## Bug

The CUDA 12.8.1 + PyTorch 2.8.0 template (`runpod/pytorch:1.0.2-cu1281-torch280-ubuntu2404`) installs **PyTorch 2.4.1** instead of 2.8.0.

## Root Cause

In `official-templates/pytorch/docker-bake.hcl` line 27:

```hcl
{ cuda_version = "12.8.1", torch = "2.8.0", whl_src = "128" },
```

The Dockerfile runs:
```
pip install torch==2.8.0 --index-url https://download.pytorch.org/whl/cu128
```

**PyTorch 2.8.0 stable wheels are not published for cu128.** Pip silently falls back to PyTorch 2.4.1+cu124.

## Evidence

Deployed a 2x B200 pod using the "Runpod Pytorch 2.8.0" template:

```
$ nvidia-smi --query-gpu=name,driver_version --format=csv,noheader
NVIDIA B200, 580.126.09
NVIDIA B200, 580.126.09

$ python3 -c "import torch; print(torch.__version__, torch.version.cuda)"
2.4.1+cu124 12.4

$ python3 -c "import torch; print(torch.cuda.get_device_properties(0))"
UserWarning: NVIDIA B200 with CUDA capability sm_100 is not compatible
with the current PyTorch installation. The current PyTorch install
supports CUDA capabilities sm_50 sm_60 sm_70 sm_75 sm_80 sm_86 sm_90.
```

Pod hostname: `7ab25c9e0ebb`
Pod IP: `38.80.152.146:31039`

## Impact

- B200 GPUs (compute capability sm_100) are **completely unusable** with PyTorch 2.4.1
- Users pay for B200 GPU time they cannot use
- RunPod's own deployment page warns "B200s only support Pytorch 2.8 and above" but the template labeled 2.8.0 doesn't deliver it

## Comparison with Other Templates

The older templates embed the actual torch version in the image tag and work correctly:

```
Pytorch 2.1: runpod/pytorch:2.1.0-py3.10-cuda11.8.0-devel-ubuntu22.04  ✅
Pytorch 2.2: runpod/pytorch:2.2.0-py3.10-cuda12.1.1-devel-ubuntu22.04  ✅
Pytorch 2.4: runpod/pytorch:2.4.0-py3.11-cuda12.4.1-devel-ubuntu22.04  ✅
Pytorch 2.8: runpod/pytorch:1.0.2-cu1281-torch280-ubuntu2404            ❌ installs 2.4.1
```

## Suggested Fix

Change the wheel source for torch 2.8.0 in `docker-bake.hcl`:

```hcl
# Current (broken):
{ cuda_version = "12.8.1", torch = "2.8.0", whl_src = "128" },

# Fix option 1 — use cu124 wheels:
{ cuda_version = "12.8.1", torch = "2.8.0", whl_src = "124" },

# Fix option 2 — use nightly wheels (as the original commit abfb7ab did):
# TORCH = "torch torchvision torchaudio --index-url https://download.pytorch.org/whl/nightly/cu128"
```

Also affects the CUDA 12.9.0 and 13.0.0 rows if the same wheel source issue applies.

## Related

- RunPod support ticket #35526 (same user, same template)
- The same `docker-bake.hcl` rows for torch 2.6.0 and 2.7.1 may also be affected — worth validating all cu128 wheel availability

dentity007:
## Summary

Adds a build-time verification step to the PyTorch Dockerfile that fails the build if the installed torch version does not match the requested version. This prevents pip from silently falling back to older PyTorch versions when wheels are missing from a CUDA-specific index.

## Problem

The current Dockerfile installs PyTorch via:

```dockerfile
RUN python -m pip install ... ${TORCH} --index-url https://download.pytorch.org/whl/cu${WHEEL_SRC}
```

When wheels don't exist at the specified index (e.g. `cu128` wheels for torch 2.8.0 were not published at the time of the last image build), **pip silently installs an older version** instead of failing. This caused:

- The `runpod/pytorch:1.0.2-cu1281-torch280-ubuntu2404` template (labeled "PyTorch 2.8.0") to actually ship **PyTorch 2.4.1+cu124**
- B200 GPUs (compute capability sm_100) to be **completely unusable** since PyTorch 2.4.1 only supports up to sm_90
- Users paying for GPU time they cannot use

## Evidence

Three separate GitHub issues report problems with the same PyTorch 2.8.0 template:

- **#114**: PyTorch 2.8.0 cu128 template installs 2.4.1 (root cause analysis with `docker-bake.hcl` line reference)
- **#98** (Aug 2025): Same template fails to start with `unsatisfied condition: cuda>=12.8`
- **#101** (Oct 2025): Same template has SSH errors and broken vLLM downloads across A40 and RTX 5090

Additionally, two RunPod support tickets (#35283, #35526) report related template issues. Support responses have attributed the problem to user-side CUDA configuration or datacenter hardware variance, but the root cause is in the build pipeline.

## Fix

A post-install `RUN` step that:

1. Extracts the expected torch version from the `TORCH` build arg
2. Checks the actually installed version via `import torch`
3. **Fails the build with a clear error message** if they don't match

```dockerfile
RUN EXPECTED_TORCH=$(echo "${TORCH}" | grep -oP 'torch==\K[0-9]+\.[0-9]+\.[0-9]+') && \
    INSTALLED_TORCH=$(python -c "import torch; print(torch.__version__.split('+')[0])") && \
    if [ "$EXPECTED_TORCH" != "$INSTALLED_TORCH" ]; then \
        echo "ERROR: PyTorch version mismatch!" >&2; \
        echo "  Expected: ${EXPECTED_TORCH}" >&2; \
        echo "  Installed: ${INSTALLED_TORCH}" >&2; \
        echo "  Wheel source: cu${WHEEL_SRC}" >&2; \
        echo "  Check that wheels exist at: https://download.pytorch.org/whl/cu${WHEEL_SRC}/torch/" >&2; \
        exit 1; \
    fi
```

## Current wheel availability (verified April 2, 2026)

| torch | cu124 | cu126 | cu128 | cu129 | cu130 |
|-------|-------|-------|-------|-------|-------|
| 2.8.0 | No | Yes | Yes | Yes | No |
| 2.9.0 | - | - | Yes | Yes | Yes |
| 2.9.1 | - | - | Yes | Yes | Yes |

The cu128 wheels for torch 2.8.0 **now exist** on the stable index. The `docker-bake.hcl` configuration is correct today, but the published Docker images were built when those wheels were missing. A rebuild with this verification step will both fix the current images and prevent this class of bug from recurring.

## Notes

- Zero additional build time for correct configurations (verification is instant)
- For broken configurations, the build fails fast with an actionable error instead of publishing a silently broken image
- The `docker-bake.hcl` does not need changes since the cu128 wheels now exist; a rebuild is sufficient
- The cu130/torch 2.8.0 row correctly uses `whl_src = "129"` since cu130 wheels don't exist for 2.8.0

Fixes #114
Related: #98, #101

### DIFF
--- a/official-templates/pytorch/Dockerfile
+++ b/official-templates/pytorch/Dockerfile
@@ -5,3 +5,18 @@ ARG WHEEL_SRC
 ARG TORCH
 
 RUN python -m pip install --resume-retries 3 --no-cache-dir --upgrade ${TORCH} --index-url https://download.pytorch.org/whl/cu${WHEEL_SRC}
+
+# Verify that the installed torch version matches what was requested.
+# pip silently falls back to older versions when wheels are missing from an index,
+# which has caused B200 GPUs to be unusable (sm_100 not supported by older PyTorch).
+# See: https://github.com/runpod/containers/issues/114
+RUN EXPECTED_TORCH=$(echo "${TORCH}" | grep -oP 'torch==\K[0-9]+\.[0-9]+\.[0-9]+') && \
+    INSTALLED_TORCH=$(python -c "import torch; print(torch.__version__.split('+')[0])") && \
+    if [ "$EXPECTED_TORCH" != "$INSTALLED_TORCH" ]; then \
+        echo "ERROR: PyTorch version mismatch!" >&2; \
+        echo "  Expected: ${EXPECTED_TORCH}" >&2; \
+        echo "  Installed: ${INSTALLED_TORCH}" >&2; \
+        echo "  Wheel source: cu${WHEEL_SRC}" >&2; \
+        echo "  Check that wheels exist at: https://download.pytorch.org/whl/cu${WHEEL_SRC}/torch/" >&2; \
+        exit 1; \
+    fi


max4c:
Closing in favor of a new PR that includes both the original verification logic and additional hardening (explicit error messages, non-empty guards, word-boundary anchoring). @dentity007 your original commit is preserved — thank you for the excellent contribution and root cause analysis!

dentity007:
Glad to see this picked up in PR #116 with the additional error handling hardening. Thanks to @max4c for expanding on the original fix and to Roman for the credit.

For anyone landing here from support tickets #35283 or #35526: the root cause (silent pip fallback when cu128 wheels are missing) is being addressed in #116. The parameter-golf:latest template regression (FA3 step time 2.5-3x slower) was a related but separate issue that has been forwarded to the DevRel team.

PR #122, opened 2026-05-13
feat: add linting, vulnerability scanning, and rework build triggers

Chmokachka:
## Summary

This PR introduces CI quality gates (linting, vulnerability scanning), reworks
the build workflow triggers, and fixes a couple of issues uncovered along
the way.

### Linting

* **ShellCheck** — lints all shell scripts in the repo. Findings are surfaced
  as inline annotations on the **Files changed** tab of the PR.
* **Hadolint** — lints Dockerfiles. Wired up in two flavors:
  * **PR workflow**: runs on every PR, findings show as inline annotations
    on the **Files changed** tab so reviewers see them next to the code.
  * **Push workflow**: runs on merges to `main`, uploads SARIF to GitHub,
    findings are viewable under **Security & quality → Code scanning**.

### Vulnerability scanning

* **Grype** image scanner added to the build pipeline. Trivy was the initial
  choice but we switched to Grype after the concerns raised in review about
  Trivy's recent security issues.
* The scan reads the bake metadata produced by the build step, extracts
  the list of built image tags, and runs Grype on each one with
  `--only-fixed --scope all-layers --fail-on high`.
* **Report-only for now.** Grype currently **does not block the push and
  does not fail the job** even when HIGH/CRITICAL fixed vulnerabilities are
  found. The action collects the affected refs into a `failed` list and
  emits a `::error::` summary in the log, but the final `exit 1` is
  intentionally commented out (with a `TODO`). Images are still pushed by
  `docker/bake-action` (`push: true`) before the scan runs, so the scan is
  currently informational only.
* Once the existing HIGH/CRITICAL findings are remediated (tracked in #124),
  we'll flip the switch so any new HIGH/CRITICAL vulnerability fails the
  job. 

### Build workflow triggers

* The build jobs now run on:
  * PR opened and subsequent pushes to the PR branch
  * Merges to `main`
  * Manual dispatch (`workflow_dispatch`)

### Fixes

* Fixed the `pip install` step in the base Dockerfile — the previous
  invocation was failing because the `get-pip.py` URL did not support
  Python 3.9.
* Exposed `max-parallelism` on `useblacksmith/setup-docker-builder` and
  tuned it down for the pytorch matrix. Building all 25 pytorch images in
  parallel was saturating the runner's network — pip downloads of multi-GB
  CUDA wheels were timing out and DNS resolution itself was failing
  (`Temporary failure in name resolution`). Limiting parallelism keeps the
  network stable while still building everything in one job.


kodxana:
The linting/scanning direction looks good, but I think this should stay out of merge until
  the scanner semantics match the PR description.

  Right now the build jobs still run `docker/bake-action` with `push: true`, then extract refs and run Trivy afterward.
  That means images are already published before the scan result is known. On top of that, `.github/actions/trivy/
  action.yml` collects failed refs but leaves `exit 1` commented out, then still prints “All images clean...”.

  So this is currently report-only scanning, not “scan before push” or a release gate. I’d either:
  - make the PR/body/workflow explicitly say this is report-only for now, or
  - change the flow so images are built/loaded locally, scanned, and only pushed after Trivy passes, with Trivy
  returning non-zero on HIGH/CRITICAL findings.

  Also worth resolving the Blacksmith OOM/build stability before merge, since this workflow becomes the foundation for
  the follow-up PRs.

ssube-runpod:
Are we comfortable introducing Trivy considering their recent security problems?

TimPietruskyRunPod (dismissed):
**Changes requested — missing `if:` guards on the Trivy steps will break unrelated PRs**

In `.github/workflows/base.yml`, the `build-autoresearch` and `build-pytorch` jobs guard the **Build** step with
`if: github.event_name == 'workflow_dispatch' || steps.changes.outputs.<x>_any_changed == 'true'`,
but the following **Extract image refs** and **Trivy scan** steps have no guard.

On any PR that doesn't touch autoresearch/pytorch, the build step is skipped, so `steps.build.outputs.metadata` is empty. The Trivy action then resolves zero image refs and hits its `exit 1` ("No image refs to scan"), failing the job even though nothing relevant changed. This isn't caught by this PR's own CI because the run that went green did have changes/dispatch.

Please add the same `if:` condition to the `Extract image refs` and `Trivy scan` steps in both jobs. (`build-base` is fine, it always builds.)

Everything else here is non-blocking from my side.

TimPietruskyRunPod (commented):
Pinning the `if:`-guard request from my Changes-Requested review to the exact lines — see inline.

.github/workflows/base.yml
```diff
@@ -86,24 +101,36 @@ jobs:
             official-templates/autoresearch/docker-bake.hcl
           push: true
 
+      - name: Extract image refs
```
TimPietruskyRunPod, line 104: These `Extract image refs` / `Trivy scan` steps have no `if:` guard, but `Build autoresearch images` above does. When the build is skipped, `steps.build.outputs.metadata` is empty and the `image-name` action runs `echo "" | jq` under `set -euo pipefail`, which exits non-zero and fails the job. Add the same `if: github.event_name == 'workflow_dispatch' || steps.changes.outputs.autoresearch_any_changed == 'true'` guard to both steps. (#124 already does exactly this.)

.github/workflows/base.yml
```diff
@@ -126,3 +155,14 @@ jobs:
             official-templates/shared/versions.hcl
             official-templates/pytorch/docker-bake.hcl
           push: true
+
+      - name: Extract image refs
```
TimPietruskyRunPod, line 161: Same missing `if:` guard here for the pytorch job — guard both `Extract image refs` and `Trivy scan` with the pytorch `*_any_changed` condition so the steps don't run (and fail on empty metadata) when the build is skipped.

Chmokachka: Done, thanks for catching that

TimPietruskyRunPod (approved):
if-guards on the autoresearch/pytorch Extract-refs + scan steps are in place and verified at the current head — my only requested change. Approving.

PR #124, opened 2026-05-18
Fix: image vulnerabilities

Chmokachka:
## Summary

Drives all `runpod/*` images to a clean Grype / Hadolint scan, restructures CI so images are **scanned before publish**, and adds several stability/perf fixes to the build pipeline. Targets every image we ship out of `official-templates/` and `helper-templates/`.

### Image vulnerabilities (Grype `--fail-on high`, `--only-fixed`)

- **base** — bumped `jupyterlab`, `notebook`, OpenSSH-related deps; pinned versions via a new `requirements.txt`. Stripped the `efa_metrics` directory from NVIDIA Nsight Compute. That directory ships an internal Go binary (`nic_sampler`) that NVIDIA builds with an old Go toolchain and was triggering recurring Go-stdlib HIGH/CRITICAL findings on every rebuild. The plugin is AWS-EFA-only (x86, AWS hardware) and never runs on RunPod, so deleting it is safe and the `find ... || true` guard keeps it a no-op on ROCm / CPU images.
- **autoresearch** — Hadolint findings fixed; aligned with new base.
- **pytorch** — Hadolint fixes.
- **rocm** — addressed all fixable CVEs; pinned the relevant deps. The `rocm644-ubuntu2204-pytorch251` variant is removed from the build matrix.
- **nvidia-pytorch** — patched OS-package CVEs; added `scrub-stale-metadata.py` (see below) to remove orphan `.dist-info` / `.egg-info` trees that kept the scanner reporting fixed wheels as still-vulnerable.

### Hadolint

- All `DL3008` / `DL3009` / `DL3015` findings fixed across the touched Dockerfiles (`--no-install-recommends`, `apt-get clean && rm -rf /var/lib/apt/lists/*`, version pins where reasonable).
- `hadolint-push` workflow now ignores the rules already accepted project-wide (matches the PR check behaviour).

### CI: scan-before-publish (addresses [#122 (review)](https://github.com/runpod/containers/pull/122))

Reworked all build workflows (`base.yml`, `nvidia.yml`, `rocm.yml`) into:
docker buildx bake (load: true, push: false) → Extract refs → Grype scan → docker-push composite action

Grype is intentionally still report-only — flipping `--fail-on` to a non-zero exit later turns this into a real gate with no further restructuring.

### CI: stability & throughput (`load: true` was risky on big matrices)

`load: true` materializes every bake target into the local Docker daemon before push, which on the pytorch matrix (~25 multi-GB CUDA/torch images) was at risk of ENOSPC and was making the "loading layer" step slow. Mitigations:

- **Enabled `containerd-snapshotter`** in `.github/actions/docker-setup`. With the default `overlay2`, BuildKit has to re-import every layer into the daemon's image store after the build. With `containerd-snapshotter` the daemon and BuildKit share one image store, so the "loading layer" step is effectively free. Verified active in a follow-up check that fails the job if the snapshotter isn't picked up.
- **pytorch Dockerfile**: switched `pip install` to a BuildKit cache mount (`--mount=type=cache,target=/root/.cache/pip,sharing=locked`) and dropped `--no-cache-dir`. Parallel matrix targets (5 CUDA × Ubuntu variants) no longer fetch the same torch / cuDNN / cuBLAS wheels independently — they share the wheel cache, and it survives across CI runs via sticky-disk. Resolves the `IncompleteRead` failures we were seeing on pytorch.org under bandwidth contention.
- **autoresearch Dockerfile**: same shape, but for `uv` — added `--mount=type=cache,target=/root/.cache/uv,sharing=locked` and bumped `UV_HTTP_TIMEOUT` from the 30s default to 1000s. Resolves the `Failed to download torch==2.9.1+cu128 ... network timeout` failures.
- Larger runners (`blacksmith-32vcpu`) and bumped `timeout-minutes` on the pytorch matrix, which was previously being killed before completion.

### CI: misc

- Upgraded GitHub Actions versions across `nvidia.yml`, `rocm.yml`, `hadolint-pr.yml`, `hadolint-push.yml`.
- Fixed `docker/setup-qemu-action` invocation that broke after the action's input rename.
- Pinned `RUNPODCTL_VERSION=v2.3.0` in `base/Dockerfile` to stop tracking `latest`.
- `autoresearch` / `pytorch` push steps are correctly gated on `steps.changes` so they only run when their template directories actually changed.

### New: `scripts/scrub-stale-metadata.py`

Small helper invoked by Dockerfiles after `pip install`. NGC base images bundle several Python packages as in-tree source builds whose `.egg-info` lives next to the source. `pip install --upgrade` upgrades the wheel install but cannot reach those bundled trees, so the scanner keeps reporting the old version even though the runtime resolves to the new one. The script reads our pinned `requirements.txt` and deletes any `.dist-info` / `.egg-info` whose `Version:` disagrees with the pin.

## What's NOT fixed (deliberate)
Two classes of findings remain across the matrix — both blocked on upstream work, neither actionable from this PR:
### 1. `filebrowser` Go transitive deps — affects **every image** that ships filebrowser
The filebrowser binary (installed via the official `get.sh`) is statically linked against vulnerable versions of `golang.org/x/crypto`, `golang.org/x/net`, and `golang.org/x/image`. Grype reports 13 HIGH/CRITICAL Go advisories for it on every rebuild:
| ID | Severity | Module | Installed | Fixed in |
| --- | --- | --- | --- | --- |
| GO-2026-5005 / 5006 / 5013 / 5017 / 5018 / 5019 / 5020 / 5021 / 5023 | Critical / High | `golang.org/x/crypto` | v0.50.0 | v0.52.0 |
| GO-2026-4918 / 5026 | High / Critical | `golang.org/x/net` | v0.52.0 | v0.53.0 / v0.55.0 |
| GO-2026-5032 | High | `golang.org/x/image` | v0.39.0 | v0.41.0 |

These can only be cleared by rebuilding filebrowser itself against bumped Go modules. The dependency bumps are already queued in the renovate batch [filebrowser#5946](https://github.com/filebrowser/filebrowser/pull/5946). Our images will inherit the fix automatically on the next filebrowser release — no change needed on our side.
### 2. `autoresearch` Python transitive deps — affects `runpod/autoresearch:*` (both ubuntu2204 and ubuntu2404)
| Package | Installed | Fixed in | Severity | Advisory |
| --- | --- | --- | --- | --- |
| `pillow` | 12.1.1 | 12.2.0 | High | GHSA-pwv6-vv43-88gr |
| `pillow` | 12.1.1 | 12.2.0 | High | GHSA-whj4-6x5x-4v2j |
| `urllib3` | 2.6.3 | 2.7.0 | High | GHSA-mf9v-mfxr-j63j |
| `urllib3` | 2.6.3 | 2.7.0 | High | GHSA-qccp-gfcp-xxvc |

These come from autoresearch's own `pyproject.toml` / `uv.lock` and need an app-level dependency upgrade in the autoresearch repo. Out of scope for this PR; tracked separately as a follow-up task.
Everything else is now clean.

## Follow-ups (separate PRs)
- Flip Grype to fail-on for HIGH/CRITICAL with fixed versions, turning the current scan ordering into a hard gate.
- Bump filebrowser version once [filebrowser#5946](https://github.com/filebrowser/filebrowser/pull/5946) is released — clears all 13 Go advisories above.
- Open autoresearch-side PR to upgrade `pillow` and `urllib3`.


kodxana:
Good vulnerability cleanup overall, especially pinning versions and removing stale metadata that causes false-positive
  Trivy reports.

  My blocker is that this PR still depends on the report-only Trivy behavior from #122. The action still does not fail
  when HIGH/CRITICAL findings are found, and the workflows still scan after `push: true`, so the CI does not prove that
  vulnerable images are blocked from publication.

  Because this PR’s goal is “fix image vulnerabilities”, I’d like to see one of these before merge:
  - Trivy exits non-zero for HIGH/CRITICAL fixed vulnerabilities and runs before publish, or
  - the PR clearly states that CI is not enforcing this yet and includes links/logs showing the claimed clean scans for
  each targeted image.

  The `skip-files` addition seems reasonable for the known civetweb cert false positives, but it makes the enforcement
  story even more important so real findings do not get hidden behind a passing workflow.

TimPietruskyRunPod (commented):
_Superseded — see the Changes Requested review below._

Correction: my earlier note here claimed `nvidia-pytorch` "now ships Jupyter." That was wrong — the image always had Jupyter; dropping `RP_SKIP_JUPYTER` just lets RunPod manage/patch it. Please disregard this review.

official-templates/base/docker-bake.hcl
```diff
@@ -14,6 +14,8 @@ target "common-base" {
     scripts = "container-template"
     proxy   = "container-template/proxy"
     logo    = "container-template"
+    requirements = "official-templates/base"
+    scrab_stale_metadata = "scripts"
```
TimPietruskyRunPod: Typo: `scrab_stale_metadata` → `scrub_stale_metadata`. It's consistent across the bake files and the `COPY --from=` in the Dockerfile, so it builds, but it reads as a mistake. Same typo in `official-templates/nvidia-pytorch/docker-bake.hcl` and `official-templates/rocm/docker-bake.hcl`.

official-templates/base/Dockerfile
```diff
@@ -85,14 +92,14 @@ RUN [[ -n $RP_SKIP_PYTHON ]] && exit 0; \
 
 COPY --from=ghcr.io/astral-sh/uv:latest /uv /uvx /bin/
 
+COPY --from=requirements requirements.txt /requirements.txt
+COPY --from=scrab_stale_metadata scrub-stale-metadata.py /tmp/scrub-stale-metadata.py
```
TimPietruskyRunPod: Same `scrab_stale_metadata` typo here in the `COPY --from=` (should be `scrub_stale_metadata`).

scripts/scrub-stale-metadata.py
```diff
@@ -0,0 +1,78 @@
...
+    for raw in pathlib.Path(requirements_path).read_text().splitlines():
+        line = raw.split("#", 1)[0].strip()
+        if "==" not in line:
+            continue
+        name, version = line.split("==", 1)
+        pinned[canonical(name)] = version.strip()
+    return pinned
+
+
+def read_meta(meta_dir: pathlib.Path) -> tuple[str, str] | None:
```
TimPietruskyRunPod, line 39: This uses PEP 604 syntax (`tuple[str, str] | None`) in an annotation evaluated at import time, so it raises on Python 3.9. It runs under the 3.12 base symlink / NGC python today so it's fine in practice — but please add `from __future__ import annotations` (or a comment pinning the assumption) since the script is generic. Also `Iterator` is referenced in the annotation on line 54 but never imported (harmless only because it's a string annotation).

official-templates/base/requirements.txt
```diff
@@ -0,0 +1,5 @@
+hf_transfer
```
TimPietruskyRunPod: Nit: trailing space after `hf_transfer`, and the file has no trailing newline (line 5). Same in the nvidia-pytorch / rocm requirements files.

TimPietruskyRunPod (changes requested):
Requesting changes — a few quality fixes (details inline). The stale-metadata approach, the efa_metrics strip, and the version pins all look good, and this PR also resolves the missing `if:` guard I flagged on #122.

- `scrab_stale_metadata` typo (bake files + `COPY --from=`).
- `scrub-stale-metadata.py` uses 3.10+ annotation syntax + an unimported `Iterator`.
- Trailing space / missing newline in the new `requirements.txt` files.

official-templates/base/docker-bake.hcl
Chmokachka: done

official-templates/base/Dockerfile
Chmokachka: done

scripts/scrub-stale-metadata.py
Chmokachka: done

official-templates/base/requirements.txt
Chmokachka: done

TimPietruskyRunPod (approved):
All four requested changes are in and verified (scrub typo across the bake files + Dockerfile, `from __future__ import annotations` + proper `Iterator` import in scrub-stale-metadata.py, and the requirements.txt whitespace/newline fixes). Threads resolved. Thanks!

TimPietruskyRunPod (commented):
Re-reviewed the new commits added since my approval (`feat: push after grype` + `feat: pytorch timeout-minutes`). Summary:

**Good — this is a real improvement.** Build now runs with `load: true, push: false`, then Extract refs → Grype → the new `docker-push` action, consistently across base/nvidia/rocm. So images are scanned **before** publish — the "scan before push" gap from the original review is now structurally closed. The autoresearch/pytorch push steps are correctly gated by `steps.changes` (same job that owns the `changes` step), and the `docker-push` composite action is sound (`set -euo pipefail`, fails on empty refs).

**One concern to confirm before merge (inline):** `load: true` materializes every image in the bake matrix into the runner's local Docker daemon before pushing. build-nvidia (single image) passed, but base and pytorch build large matrices of multi-GB CUDA/torch images — loading them all at once risks exhausting runner disk (ENOSPC). The current run is the first with `load: true` on those matrices and build-base/build-rocm are still in progress; worth watching them specifically. The `blacksmith-32vcpu` + `timeout-minutes: 240` bump helps CPU/time but not disk.

**Note:** grype is still report-only by design, so push isn't actually blocked on findings yet — but the ordering means flipping the `exit 1` later turns it into a real gate with no further restructuring. Fine as a foundation.

No blocking objection from me on the design; just confirm the base/pytorch runs go green with `load: true`.

.github/workflows/base.yml
```diff
@@ -46,7 +46,8 @@ jobs:
           files: |
             official-templates/shared/versions.hcl
             official-templates/base/docker-bake.hcl
-          push: true
+          load: true
```
TimPietruskyRunPod, line 49: `load: true` imports the entire bake matrix into the runner's local Docker daemon before the push step. For the base CUDA matrix (and especially the pytorch matrix of ~25 multi-GB CUDA/torch images) this can blow the runner's disk (ENOSPC) since all images coexist locally instead of streaming straight to the registry. nvidia (1 image) is fine; please confirm build-base/build-pytorch actually pass with this. If disk becomes the bottleneck, consider load→scan→push→`docker image rm` per ref (or per small batch) so they don't all pile up at once.

Chmokachka: I'm still experimenting with `load: true`, as it really blows the runner's disk. Moving the PR back to draft until I'm done with it.

TimPietruskyRunPod (approved):
Re-reviewed at the current head after the rebase onto main + push-after-grype rework. All build jobs are green with `load: true` (base ~1h40m, pytorch ~5h, rocm ~2h30m), so the disk concern from my earlier comment did not materialize — resolving it. Note pytorch is brushing ~5h, so the matrix is slow under `load: true`, but it passes. Approving.

PR #125, opened 2026-05-22
fix: sonarqube findings

Chmokachka:
## Summary

Addresses all SonarQube findings on the active codebase except one **Security Hotspot** ("container running as root") which is tracked separately. Also cleans up long-archived templates and bumps `RELEASE_VERSION` to `1.0.6`.

### Shell-script hardening (Bash & SonarQube `bash:*`)

- `[ ... ]` → `[[ ... ]]` everywhere (`bake.sh`, `container-template/start.sh`, `scripts/update-template-readme.sh`, `official-templates/*/pre_start.sh`, `official-templates/*/Dockerfile`).
- Quote every variable expansion (`$VAR` → `"$VAR"`) so paths with spaces / unset vars don't silently misbehave (`start.sh` SSH-host-key block, Jupyter token, etc.).
- Error messages now go to **stderr** (`>&2`) in `scripts/update-template-readme.sh`.
- `pytorch/Dockerfile`: enable `SHELL ["/bin/bash", "-o", "pipefail", "-c"]` and split `TORCH="torch==X torchvision==Y ..."` into a real array with `read -ra TORCH_PKGS <<<` so each pin reaches `pip` as a separate argument.

### Supply-chain pinning + integrity checks (`docker:*`, GH Actions)

- **Third-party GitHub Actions pinned to commit SHA** (was floating tags):
  - `docker/bake-action` → `6614cfa25e...`
  - `reviewdog/action-shellcheck` → `4c074582...`
  - `SonarSource/sonarqube-scan-action` → `7006c449...`
  - `sonarsource/sonarqube-quality-gate-action` → `cf038b0e...`
- **Downloads pinned + SHA-256 verified**, no more `curl | bash`:
  - `get-pip.py`: pinned to commit `0d4515b4`; separate hashed copy for Python 3.9 (`pip/3.9/get-pip.py`).
  - `filebrowser`: pinned to `v2.63.5`, replaces the upstream `curl … | bash` installer with verified tarball.
  - `runpodctl` (autoresearch): pinned `v2.3.0`, SHA-256 verified.
  - Trivy installer hardened with `--proto '=https' --tlsv1.2`.
- `base/requirements.txt`: pin `hf_transfer`, `ipywidgets`, `jupyter-archive`.
- `autoresearch/Dockerfile`: `uv sync --frozen --no-build` so no source-distribution `setup.py` runs at build time.

### GitHub Actions injection hardening (`github-actions:*`)

- `.github/actions/update-readme/action.yml` no longer interpolates `${{ inputs.* }}` directly into shell scripts. Inputs are passed as `env:` vars (`TEMPLATE_PATH`, `TEMPLATE_ID`) and referenced as `"${TEMPLATE_PATH}"`, etc.
- Dropped unused `packages: write` permission from `base.yml`, `nvidia.yml`, `rocm.yml`.

### HTML hardening (`Web:*`, `javascript:*`) in `container-template/proxy/readme.html`

- `<html lang="en">`.
- Showdown CDN script now has `integrity="sha512-…"`, `crossorigin="anonymous"`, `referrerpolicy="no-referrer"`.
- Inline script becomes `type="module"` so we can `await` cleanly and ditch the `.then(...)` chain.

### Welcome-banner fix

- `base/Dockerfile`: bashrc banner was using `printf "%s"` with embedded `\n`, which rendered literal `\n` in the SSH login screen. Switched to a heredoc (`cat <<'BANNER' … BANNER`) so newlines and prompt formatting render correctly.

### SonarQube setup

- New `.github/workflows/sonarqube.yml`: scan + quality-gate job on every push.
- New `sonar-project.properties` with project key and a coverage exclusion for `scripts/scrub-stale-metadata.py`.

### Cleanup

- Removed long-archived templates under `archived/` (kasm-desktop, discoart, bittensor, fast-stable-diffusion, invokeai, oobabooga, stable-diffusion-comfyui, stable-diffusion-webui, tensorflow, vs-code, vscode-server, plus archived `pytorch`). Net `-13 581 / +153` lines — these have been unused for releases and were generating SonarQube noise.
- Bumped `RELEASE_VERSION` `1.0.5` → `1.0.6`.

## Known remaining SonarQube finding

**One Security Hotspot is intentionally left for a follow-up PR:**

> _Container running as root user_

## Follow-ups

- Dedicated PR for the "root user" hotspot: design the non-root model for interactive RunPod containers (which services need privileges, how to drop them, what stays as root via `setcap` / explicit `USER root` blocks).


kodxana:
There are several good hardening changes here: moving action inputs through env vars, verifying downloaded artifacts,
  dropping unused permissions, and pinning some third-party actions.

  A few things I’d fix before merge:

  1. The PR says third-party actions are pinned to commit SHAs, but several are still floating tags, including

  3. `git diff --check pr-124..pr-125` reports whitespace issues in the touched files. Not functionally huge, but worth
  cleaning up while we’re touching Sonar/lint hygiene.

  The archived-template deletion is large and probably fine if intentional, but it makes this PR much harder to audit
  alongside security changes.

TimPietruskyRunPod (commented):
_Superseded — see the Changes Requested review below._

(Removed an earlier note questioning the `archived/` deletion — that cleanup is intentional and fine.)

.github/workflows/sonarqube.yml
```diff
@@ -0,0 +1,33 @@
+name: SonarQube
+
+on:
+  push:
```
TimPietruskyRunPod, line 4: `push:` with no branch filter runs the scan + quality gate on every branch push. Suggest scoping to `branches: [main]` plus `pull_request:` to avoid noise and duplicate runs.

.github/workflows/sonarqube.yml
```diff
@@ -0,0 +1,33 @@
...
+
+permissions:
+  contents: read
+
+jobs:
+  sonarqube:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v5
```
TimPietruskyRunPod: The PR description says third-party actions are pinned to commit SHAs, but this is still a floating tag — and it's `@v5` here while the other workflows use `actions/checkout@v6`. The same floating-tag issue remains for `hadolint/hadolint-action@v3.1.0`, `github/codeql-action/upload-sarif@v4`, `docker/login-action@v4`, `docker/setup-qemu-action@v4`, and `useblacksmith/setup-docker-builder@v1`. Pin them to SHAs or adjust the description. (@kodxana raised this too.)

TimPietruskyRunPod (changes requested):
Requesting changes (details inline). Shell hardening (`[[ ]]`, quoted expansions, passing inputs via `env:` to avoid `${{ }}` shell injection) and the verified/pinned downloads all look good.

- PR says third-party actions are SHA-pinned, but several are still floating.
- `sonarqube.yml` triggers on `push:` for all branches.
- whitespace/newline nits (also noted by @kodxana).

.github/workflows/sonarqube.yml
Chmokachka: done

.github/workflows/sonarqube.yml
Chmokachka: done

TimPietruskyRunPod (approved):
Both requested changes are in and verified: third-party actions are now pinned to commit SHAs across the workflows/composite actions, and `sonarqube.yml` is scoped to `push: [main]` + `pull_request`. Thanks!

PR #126, opened 2026-05-29
Feat/TEM-7 testing

Chmokachka:


.github/workflows/base.yml
```diff
@@ -168,4 +187,17 @@ jobs:
...
+      - name: Smoke test
+        if: github.event_name == 'workflow_dispatch' || steps.changes.outputs.pytorch_any_changed == 'true'
+        uses: ./.github/actions/smoke-test
+        with:
+          image-refs: ${{ steps.refs.outputs.refs }}
+          profile: gpu
+          runpod-api-key: ${{ secrets.TESTING_RUNPOD_API_KEY }}
+          ssh-private-key: ${{ secrets.TESTING_RUNPOD_SSH_PRIVATE_KEY }}
+          test-jupyter: true
+          exclude_instances: |
```
TimPietruskyRunPod: Input name mismatch: the action declares this input as `exclude-instances` (hyphen, consumed at `.github/actions/smoke-test/action.yml` as `inputs.exclude-instances`), but here it's passed as `exclude_instances` (underscore). GitHub will treat it as an unexpected input and ignore it, so the `*Blackwell*` exclusion never applies — which is exactly the case this is meant to prevent (PyTorch ≤2.6 wheels have no kernels for sm_100/sm_120, so a pod landing on a Blackwell host fails with "no kernel image is available for execution on the device"). Rename to `exclude-instances`.

TimPietruskyRunPod (changes requested):
Requesting one change (inline). The smoke-test framework looks solid overall — nice that the functional check is derived from the image ref rather than the profile, and that the pod is always terminated.

One real bug: the pytorch job's `exclude_instances` input name doesn't match the action, so the `*Blackwell*` exclusion is silently dropped.

Non-blocking, worth a look:
- The smoke step needs `secrets.TESTING_RUNPOD_API_KEY` / `TESTING_RUNPOD_SSH_PRIVATE_KEY`, which won't be present on fork PRs — the step will fail for external contributors.
- Every qualifying PR now boots real (paid) RunPod pods; fine if that's intended.
- Blacksmith still reported a `build-pytorch` OOM on this branch.

.github/workflows/base.yml
```diff
@@ -113,11 +137,32 @@ jobs:
...
+    needs: build-autoresearch
+    if: needs.build-autoresearch.result == 'success'
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+
+      - name: Smoke test
+        if: github.event_name == 'workflow_dispatch' || steps.changes.outputs.autoresearch_any_changed == 'true'
```
TimPietruskyRunPod: This `if:` references `steps.changes.outputs.autoresearch_any_changed`, but `test-autoresearch` has no `Check if build is needed` step — the `id: changes` step lives in `build-autoresearch`. In this job the expression resolves to an empty string, so the condition is `false` on `pull_request`: the smoke test runs only on `workflow_dispatch` and is silently skipped on PRs that actually change autoresearch. Either expose the changed flag as a `build-autoresearch` job `output` and gate this job at the job level (`needs.build-autoresearch.outputs.*`), or move the `changes` detection into this job.

.github/workflows/base.yml
```diff
@@ -168,4 +213,28 @@ jobs:
...
+    needs: build-pytorch
+    if: needs.build-pytorch.result == 'success'
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+
+      - name: Smoke test
+        if: github.event_name == 'workflow_dispatch' || steps.changes.outputs.pytorch_any_changed == 'true'
```
TimPietruskyRunPod: Same issue as `test-autoresearch`: `steps.changes.outputs.pytorch_any_changed` doesn't exist in `test-pytorch` (no `changes` step here), so the pytorch smoke test is skipped on PRs that change pytorch and only runs on `workflow_dispatch`. Gate via a `build-pytorch` job output instead.

TimPietruskyRunPod (changes requested):
Re-reviewed after the test-job split. The `exclude-instances` typo is fixed ✅ (resolved that thread). But moving the smoke test into separate `test-*` jobs introduced a new issue (inline): for autoresearch and pytorch the per-step `if:` still references `steps.changes.*`, which doesn't exist in the new test jobs — so those smoke tests will be silently skipped on the PRs that actually change them.

Minor: the new `test-*` jobs use `actions/checkout@v6` (floating), whereas #125 just pinned every checkout to a SHA — worth keeping consistent.

.github/workflows/base.yml
```diff
@@ -126,12 +150,32 @@ jobs:
         with:
           image-refs: ${{ steps.refs.outputs.refs }}
 
+  test-autoresearch:
+    runs-on: blacksmith-4vcpu-ubuntu-2204
+    needs: build-autoresearch
+    if: needs.build-autoresearch.result == 'success'
```
TimPietruskyRunPod: `build-autoresearch` runs on every PR (its job-level `if` is `always()`); only its *steps* are gated by `steps.changes`. So when autoresearch is unchanged, the build/refs steps skip and `build-autoresearch.outputs.image-refs` is empty — but this job still runs, because skipped steps leave the job `result == 'success'`. The Smoke step then invokes the action with empty refs, and `generate_test_manifest.py` returns 1 on a non-list/empty `--refs` (`if not isinstance(refs, list) or not refs`), failing the job. Net: every PR that doesn't change base/shared/autoresearch gets a red `test-autoresearch`. Gate this job on the build having actually run — e.g. add a `changed`/`built` output to `build-autoresearch` and include it here, or `if: needs.build-autoresearch.result == 'success' && needs.build-autoresearch.outputs.image-refs != ''`.

.github/workflows/base.yml
```diff
@@ -189,4 +233,27 @@ jobs:
...
         uses: ./.github/actions/docker-push
         with:
-          image-refs: ${{ steps.refs.outputs.refs }}
\ No newline at end of file
+          image-refs: ${{ steps.refs.outputs.refs }}
+
+  test-pytorch:
+    runs-on: blacksmith-4vcpu-ubuntu-2204
+    needs: build-pytorch
+    if: needs.build-pytorch.result == 'success'
```
TimPietruskyRunPod: Same issue as `test-autoresearch`: on a PR that doesn't change base/shared/pytorch, `build-pytorch.outputs.image-refs` is empty but this job still runs (`result == 'success'`), so the smoke test fails on empty refs. Gate on the build actually having produced refs.

TimPietruskyRunPod (changes requested):
Re-reviewed after the test-job rework. The original `steps.changes` bug is gone (good — the broken step-level `if` was removed). But the new gating flips it into the opposite failure: the `test-*` jobs now run with **empty image-refs** on any PR that doesn't touch that image, and `generate_test_manifest.py` exits 1 on empty/invalid `--refs` — so the job goes red. Details inline. (base/nvidia/rocm are unaffected since those always build; this only bites autoresearch/pytorch, which are gated by `changed-files`.)

.github/workflows/base.yml
Chmokachka: done

.github/workflows/base.yml
Chmokachka: done

TimPietruskyRunPod (approved):
The smoke-test gating is fixed correctly. `build-autoresearch`/`build-pytorch` now expose a `changed` output, and the `test-*` jobs gate on `result == 'success' && (workflow_dispatch || changed)` (test-pytorch via a per-CUDA matrix + artifact download with outcome guards). Verified in CI: `test-autoresearch` now **skips** on a no-change PR instead of failing on empty refs. `exclude-instances` is correct too. Approving — just let the in-progress build-pytorch matrix / build-rocm finish green before merging.

PR #127, opened 2026-06-08
feat: cosign image signing

Chmokachka:
Cosign image signing


.github/workflows/base.yml
```diff
@@ -21,6 +21,8 @@ permissions:
 jobs:
   build-base:
     runs-on: blacksmith-16vcpu-ubuntu-2204
+    permissions:
```
TimPietruskyRunPod, line 24: A job-level `permissions:` block *replaces* the top-level one entirely, so this job now runs with `contents: none` (the top-level `contents: read` at line 19 no longer applies here). `actions/checkout` with `fetch-depth: 0` wants `contents: read` — it works today only because the repo is public, and your own COSIGN.md §2 prescribes both. Add `contents: read` alongside `id-token: write`. Same for `build-autoresearch` (~line 105), `build-pytorch` (~line 204), and the build jobs in `nvidia.yml` / `rocm.yml`.

TimPietruskyRunPod (changes requested):
The keyless-cosign design is solid and `docs/COSIGN.md` is genuinely excellent — clear trust model, runbooks, consumer instructions. A few things before merge:

1. **Per-job `permissions:` drops `contents: read`** (inline). Not fatal here because the repo is public (CI passes), but it contradicts your own COSIGN.md §2 (which lists both `contents: read` and `id-token: write`) and would break checkout if the repo ever goes private. Cheap to make correct.
2. **Confirm the signed digest is the registry digest.** `image-name` derives digests from `containerimage.digest` in the *build* metadata, but the build runs `load: true` / `push: false` and the image is pushed later by `docker-push`. Please verify the build-time `containerimage.digest` equals the manifest digest in the registry after `docker push` — otherwise `cosign sign` is signing a digest that isn't what consumers pull.
3. **Inherited + own red CI:** this is stacked on #126, whose `test-*` jobs fail on PRs that don't touch the gated images (see my #126 review), and #127's own `test-pytorch` / `test-rocm` are currently red. Worth resolving before this comes off the stack.

.github/workflows/base.yml
Chmokachka: done

TimPietruskyRunPod (approved):
Both findings addressed: `contents: read` is now on every signing job (alongside `id-token: write`), and the digest concern is resolved empirically — all build jobs pass with cosign **sign + verify** running inline, which would fail if the signed `containerimage.digest` were not the registry digest. Full CI is green (build + per-CUDA pytorch matrix + rocm + all smoke tests). COSIGN.md is great. Approving.

PR #132, opened 2026-06-18
feat: TEM-12 - matrix builds for nvidia-pytorch & rocm + CI cost/caching optimizations

Chmokachka:
### Why

Adding a new PyTorch / ROCm version used to mean hand-writing a whole new bake target (duplicated tags + args) and, worse, every variant was built on a single runner in one bake invocation - slow, and one bad variant failed the whole batch. This PR makes versions data, not code, gives each version family its own runner (mirroring the matrix pattern already used by `pytorch` in `base.yml`), and - following concerns about runner usage and caching - trims how much CI work each PR actually triggers.

### What changed

**Declarative version lists in `docker-bake.hcl` (nvidia-pytorch + rocm)**

Combinations are now a single `variable` list, and one templated `matrix` target renders the tags + `BASE_IMAGE` from each combo's fields. Adding a version = one line in the list, no new target. Per-family groups (`pytorch2511`, `rocm644`) select their subset via an `if` filter, so a workflow shard builds exactly its slice. Example:

```hcl
variable "ROCM_TORCH_COMBINATIONS" {
  default = [
    { rocm = "6.4.4", ubuntu = "22.04", python = "3.10", torch = "2.6.0" },
    { rocm = "6.4.4", ubuntu = "24.04", python = "3.12", torch = "2.6.0" },
    { rocm = "6.4.4", ubuntu = "24.04", python = "3.12", torch = "2.7.1" },
  ]
}
```

**Matrix-per-runner in CI (`nvidia.yml`, `rocm.yml`)**

- `build-*` and `test-*` jobs gained `strategy.matrix` (`nvidia: [pytorch2511]`, `rocm: [rocm644]`) with `fail-fast: false` → each family builds in parallel on its own runner, and one family failing no longer kills the others.
- `bake-action` gets `targets: ${{ matrix.* }}` so each shard builds only its group.
- Refs handoff reworked from a job `output` to per-shard artifacts (`nvidia-refs-<matrix>`, `rocm-refs-<matrix>`); the matching `test-*` shard downloads its artifact (`continue-on-error`) and only smoke-tests when refs are present, keeping each version's build/test fully independent.

**Cut redundant CI work**

- Added `concurrency` to all three workflows (`base.yml`, `nvidia.yml`, `rocm.yml`): `group: <workflow>-<ref>` with `cancel-in-progress` only for PRs. A series of commits on a PR now only pays for the latest one - superseded ~2h runs are cancelled. `push` to `main` never cancels (release builds must finish).
- Per-workflow `paths:` filters keep the three workflows isolated (a change under `rocm/` doesn't trigger `nvidia.yml`/`base.yml`, etc.), and the expensive `pytorch` / `autoresearch` matrices in `base.yml` stay gated by `changed-files` so they only rebuild when their own (or `base`/`shared`) sources change. `build-base` itself always runs - `pytorch`/`autoresearch` build `FROM` the branch-suffixed base image it pushes, so skipping it would break their `FROM` pull.

**Grype / Docker setup**

- Parallel Grype scans with a configurable cap (`max-parallel-scans`, default `2`) instead of scanning one image at a time.
- Scan scope `all-layers` → `squashed`: scan s only the final image filesystem (what actually ships) - faster, and drops false positives for files not present in the shipped image.
- Cached the Grype vulnerability DB (`~/.cache/grype/db`, daily-rotated key) so it isn't re-downloaded on every scan.
- Removed `Set up QEMU` from `docker-setup`: all targets are `linux/amd64` on amd64 runners, so binfmt emulation was unused. Also eliminates the `Unable to reserve cache … binfmt …` warning from jobs racing for that key.

**Housekeeping**

- Build runners reduced `32vcpu` → `16vcpu` (`nvidia`, `rocm`, base/pytorch builds).
- Pinned `actions/checkout`, `upload-artifact`, `download-artifact`, and the new `actions/cache` (v6.1.0) to commit SHAs.
- `test-rocm` runner bumped `ubuntu-2204` → `ubuntu-2404` to match the build host.
- NVIDIA smoke test keeps `min-cuda-version: "13.0"` (NGC `pytorch:25.11` ships torch built against CUDA 13.0 and won't init on 12.x-driver hosts).


.github/workflows/base.yml
```diff
@@ -86,7 +86,7 @@ jobs:
     if: needs.build-base.result == 'success'
     steps:
       - name: Checkout
-        uses: actions/checkout@v7
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10
```
TimPietruskyRunPod: This pins to `df4cb1c` = checkout **v6**, but the build jobs (and main, after #136 merged) are on `9c091bb` = **v7**. Same in `test-pytorch` (base.yml), `test-nvidia` (nvidia.yml), and `test-rocm` (rocm.yml). Repin all of these to `9c091bb` so checkout is consistently v7 across build and test jobs.

TimPietruskyRunPod (changes requested):
Nice refactor — versions-as-data + per-family matrix mirrors the existing pytorch pattern, the artifact hand-off is the right shape, and you correctly SHA-pinned the artifact actions to the matched pair (upload `043fb46d` = v7.0.1, download `3e5f45b2` = v8.0.1). No empty-refs concern since nvidia/rocm always build.

One fix before merge: the **test-job checkouts are pinned to `df4cb1c` (v6)**, while the build jobs in these same files are on `9c091bb` (v7). That's inconsistent, and it silently reverts the v7 bump that #136 just merged for those lines. Please repin the test-job checkouts to `9c091bb` (v7). Details inline.

Nit: `official-templates/rocm/docker-bake.hcl` is missing a trailing newline.

.github/workflows/base.yml
Chmokachka: done

ssube-runpod (approved):

PR #161, opened 2026-08-25
Feat: TEM-42 ComfyUI template + smoke harness on REST API v2 + GPU × CUDA matrix

Chmokachka:
Two related changes. The ComfyUI template moves into this repo and gains a
CUDA matrix; the smoke-test harness is reworked onto REST API v2 and grown
into something that can answer "which GPUs and which CUDA versions does this
image actually run on".

## ComfyUI image

`official-templates/comfyui/` — migrated from
[runpod-workers/comfyui-base](https://github.com/runpod-workers/comfyui-base).

- Two-stage build on `ubuntu:24.04`. The bake matrix produces one image per
  CUDA/torch pairing, tagged
  `runpod/comfyui:<version>-comfyui<comfyui_version>-cuda<12.8|13.0>`.
- ComfyUI and all four bundled custom nodes (Manager, KJNodes, Civicomfy,
  RunpodDirect) are pinned by commit SHA. `scripts/fetch-hashes.sh` prints
  refreshed SHAs in HCL form for the bake file.
- `scripts/prebake-manager-cache.py` populates the ComfyUI-Manager registry
  cache at build time, so a cold start skips ~127 paginated requests to
  `api.comfy.org`.
- `scripts/start.sh` copies the baked install to `/workspace` on first boot
  and, for pods created from the old CUDA 12.4 template, runs a one-time
  migration: ComfyUI and the bundled nodes move to the pinned versions while
  models, inputs, outputs, user settings and user-installed nodes are kept.
  It also brings up FileBrowser (`:8080`), JupyterLab (`:8888`) and SSH.
- `NVIDIA_REQUIRE_CUDA=""` is set deliberately, so the container starts on
  hosts with an older 12.x driver instead of being rejected by the NVIDIA
  prestart hook.

`comfyui.yml` is a reusable build workflow in the same shape as the other
families: build → Grype → push → cosign → smoke test, matrixed over
`cuda128` / `cuda130`. `release.yml` gains the family, its path filter and
its release gate, so a release is not tagged unless ComfyUI built and tested
green.

## Test harness

`runpodctl` subprocess calls are gone, replaced by `runpod_smoke/api.py`
against REST API v2; `runpod_smoke/runpodctl.py` is deleted.

**ComfyUI checks.** Reachability on public `:8188`, and an opt-in functional
check that provisions the model in `tests/comfyui/models.json` through
ComfyUI-RunpodDirect, POSTs `tests/comfyui/workflows/gsl_starter_1_1.api.json`,
waits for completion and validates a real PNG through the public proxy. The
generated image is uploaded as an artifact.

**Readiness accepts either SSH endpoint.** RunPod does not always allocate a
public port for `22/tcp`, and when it does it can be late — a measured pod
was proxy-only at t+601s and got its port at t+700s. A pod that never gets
one exposes only `ssh.proxy`, which needs an opaque routing token and a
pseudo-terminal, so it is invoked from the API's own `ssh.proxy.command` with
a forced `-tt`. Whichever endpoint answers is reused by every later check.

Endpoints are passed **by value**, never looked up by address: every pod's
proxy answers at the same `ssh.runpod.io:22` and differs only by the token in
`username`, so anything keyed on `(host, port)` would be one global slot that
parallel pods overwrite — and a wrong-token connection does not fail, it
lands on a different live pod and returns a plausible result for it.

**CUDA axis.** `cuda_versions: all` expands into a GPU × CUDA matrix built
from the catalog's per-version capacity and pinned with
`gpu.allowedCudaVersions`, so unavailable pairings are dropped at planning
time rather than costing a pod each. When no pairing survives, the gap is
recorded as a `SKIP` row — planning that yields nothing must not read as
"nothing failed".

**Cloud tiers.** The catalog's `cudaVersions` is scoped to one tier and the
tiers do not nest — 16 of 47 GPUs are community-only, while A40, L4, H100 SXM,
B200 and B300 have no community hosts. `CLOUD_TYPE` therefore takes a list or
`ALL`, plans once per tier, and shares one worker pool.

**Who a timeout belongs to is read from evidence, not guessed.** Nothing
visible from outside separates a broken image from a slow platform: `status`
reads `RUNNING` from scheduling time onward, so a pod still downloading a
50GB image, a pod whose entrypoint cannot execute, and a pod RunPod never
routed to all look identical. The verdict therefore comes from the two log
streams, worst evidence first:

| evidence | verdict |
|---|---|
| `nvidia-container-cli` rejected the container | `FAIL` — deterministic, no retry |
| either log stream shows it could not start at all | `FAIL` — deterministic, no retry |
| system log has no `create/start container` yet | `STUCK` — still pulling |
| container log announced sshd | `STUCK` — RunPod never routed to it |
| container log empty | `STUCK` — never started |
| container started, no sshd line | `UNVERIFIED` |
| neither stream readable | `UNVERIFIED` |

Startup failures are looked for in **both** streams, and in the *unfiltered*
system log: a container that was never created has no stdout of its own, so an
OCI refusal appears only on RunPod's side. `SYS_LOG_ERROR_PATTERN` narrows
that stream for display and matches error/fail/crash wording only, so markers
like `exec: "/start.sh": no such file or directory` never reach `sys_errors` —
classifying from that subset downgraded a broken entrypoint to a capacity gap.

`UNVERIFIED` is a fourth summary bucket, counted and printed separately from
`SKIP`, and fatal except under `ON_SKIP=pass`. The point is that "we could
not tell whether this image works" is never filed under "no capacity" and
never quietly green on a release gate.

**New manual workflow.** `gpu-compatibility.yml` boots any image on every
NVIDIA GPU in the catalog and reports a per-GPU pass/fail matrix plus a JSON
artifact. It accepts several images at once, a cloud tier
(SECURE / COMMUNITY / ALL), a CUDA axis, exclude patterns, and either
`instances` (an explicit fallback chain) or `check-all-gpu` — the two are
mutually exclusive and CI fails on both or neither rather than silently
ignoring one.

**Unit tests.** `tests/unit/` covers the planning and classification logic
with stdlib `unittest` — no network, no credentials, no pods, 55 tests in
under a second. `harness-tests.yml` runs it on every PR that touches
`tests/`, plus a contract check on the manifest generator.

## Incidental fixes

- `exclude-instances: "*Blackwell*"` in `base.yml` matched nothing. The
  patterns are compared against catalog *display* names (`B200`,
  `RTX PRO 6000`); only the ids contain the word "Blackwell".
- Under `check-all-gpu` the sweep silently inherited the action's 16 GB vRAM
  floor, dropping RTX 3070 and RTX A2000 from a run that claims full
  coverage. `gpu-compatibility.yml` now passes `min-vram-gb: "0"`.
- The `-cluster` smoke test no longer needs `min-cuda-version: "12.8"` — the
  floor is parsed from the image tag.

## Hadolint

The ComfyUI Dockerfile is added to both lint matrices. Three of the four
warnings are fixed rather than suppressed, so `DL3008` is the only ignored
rule — matching `base`, `verify-nccl` and `pytorch-cluster`.

- `DL4001` (wget and curl both used) — the two `wget` calls that fetch
  `cuda-keyring` are now `curl -fSL -o`. Both packages stay installed: this
  is a user-facing dev box and customers may reach for either. Fixing this
  also clears the two `DL3047` findings.
- `DL3003` (`cd` instead of `WORKDIR`) — one `RUN` walked through five
  directories to init a git repo in each, which `WORKDIR` cannot express
  without splitting it into five layers. `WORKDIR` now points at the ComfyUI
  root and the custom nodes use `git -C custom_nodes/<name>`, so the `cd`
  chain is gone and it is still one layer.
- `DL4006` (pipefail before a pipe) — the pipe was
  `echo "$sha  file" | sha256sum -c -`, where the left side cannot fail, so
  `pipefail` would change nothing. The standard remedy, `SHELL [... -o
  pipefail ...]`, would change every later `RUN` in the image. The checksum
  is written to a file and verified from it instead, removing the pipe.

## What the compatibility sweep found

53 pods across 7 compute capabilities (sm_80 … sm_120) and both cloud tiers,
over all 11 `runpod/pytorch:1.2.0-rc.162` builds:

- The host-driver floor equals the CUDA version of the image's base layer:
  `cu1281` needs ≥ 12.8, `cu1290` ≥ 12.9, `cu1300` ≥ 13.0. Below the floor
  the CUDA forward-compatibility layer carries datacenter cards but not
  GeForce.
- **RTX 5080 + `cu1300` fails 6/6** with
  `Error 804: forward compatibility was attempted on non supported HW`, across
  two provider addresses, while the same cards run `cu1281` and `cu1290` and
  RTX 5090 on the same driver runs `cu1300`. It has to be excluded by card.

## Testing

Latest commit is green across all 26 checks, including the unit suite:
[run 34338141863](https://github.com/runpod/containers/actions/runs/34338141863).

The harness is shared by every image family, so all four were built and
smoke-tested on the reworked code:
[run 34130325852](https://github.com/runpod/containers/actions/runs/34130325852)
— `base / test-base`, `nvidia / test-nvidia`, `rocm / test-rocm` and both
`comfyui / test-comfyui` variants green.

`gpu-compatibility.yml` is dispatch-only, so GitHub shows no Run button until
it is on `main`. Both selection paths were exercised from this branch with
temporary push-triggered workflows that copied its steps verbatim, since
removed:

- explicit `instances` —
  [run 33541276473](https://github.com/runpod/containers/actions/runs/33541276473/job/99967838345)
- `check-all-gpu` on SECURE —
  [run 33541276620](https://github.com/runpod/containers/actions/runs/33541276620/job/99967838887)
- the reworked harness on two never-before-tested pytorch builds, plus a
  no-pod check that a planning gap exits non-zero —
  [run 34130320674](https://github.com/runpod/containers/actions/runs/34130320674)

`tests/README.md` documents the manifest schema, the CUDA axis, the cloud
tiers, both SSH endpoints, the timeout evidence table, every environment
variable and the outcome buckets.


kodxana:
Thanks, the ComfyUI builds look good. I found three issues in the shared harness that should be fixed before merging:

- `_SSH_ENDPOINTS` is keyed only by `(host, port)`. Two pods sharing an SSH proxy address overwrite each other’s routing username, so checks can run against the wrong pod. Please carry the pod-specific endpoint, including its username, through the checks.
- With explicit instances and a CUDA axis, unavailable matching capacity can produce no jobs and no result rows. This returns `0 PASS, 0 FAIL, 0 SKIP` with exit 0, even with `ON_SKIP=fail`. Please record the missing coverage as SKIP and apply the configured policy.
- Every readiness timeout becomes STUCK unless it matches the NVIDIA requirement-error pattern. That can hide image startup failures as infrastructure skips. Please check for container startup failures before classifying the timeout, and keep ambiguous cases explicitly unverified.

Could you add regression tests for these cases and exercise representative existing image families with the updated harness? The current run skipped base, NVIDIA and ROCm, though they all use it.

mariiachekmasova-runpod:
Fixed:

1. **`_SSH_ENDPOINTS` keyed by `(host, port)`** — fixed. The registry is gone.
   Endpoints are now an immutable `SshEndpoint(host, port, user, command, pty)`
   passed by value through every check, and `wait_for_running` returns the one
   that answered. Worth noting the failure mode was silent, not a crash: the
   proxy address is the same for all pods, so a wrong-token connection landed
   on a live pod and returned a plausible result for it.

2. **Explicit instances + CUDA axis producing no jobs and no rows** — fixed.
   `_cuda_per_version_jobs` now records the gap as a SKIP naming each
   candidate and its reason. Plus a backstop for the whole class of bug: a run
   with zero result rows exits 1 regardless of `ON_SKIP`, since nothing was
   attempted. Verified end to end against the live catalog in
   [this job](https://github.com/runpod/containers/actions/runs/34130320674/job/101768611661):
   explicit instances pinned to a CUDA version the catalog does not offer,
   asserting a non-zero exit and a SKIP row that names the reason. Creates no
   pods.

3. **Every timeout becoming STUCK** — fixed, in the order you suggested.
   Deterministic image faults are checked first (`nvidia-container-cli`, and a
   new detector for containers that could not start at all — bad entrypoint,
   wrong architecture, OOM); both are `FAIL` with no retry. The rest is read
   from the two log streams rather than guessed, because `status` says
   `RUNNING` from scheduling time onward and cannot tell a 50GB pull from a
   wedged host. Ambiguous cases are a new `UNVERIFIED` bucket — counted
   separately from `SKIP` and fatal except under `ON_SKIP=pass`, so an image we
   could not verify blocks a release instead of passing as a capacity gap.

**Regression tests** — `tests/unit/` is new; the harness had no suite at all.
52 tests, stdlib `unittest`, no network or credentials, under a second.
`harness-tests.yml` runs them on every PR touching `tests/`. Each regression
test was checked to fail with its fix reverted.

**Existing families** — [run
34130325852](https://github.com/runpod/containers/actions/runs/34130325852):
`base / test-base`, `nvidia / test-nvidia (pytorch2511)`, `rocm / test-rocm
(rocm644)` and both `comfyui / test-comfyui` variants green on the reworked
harness.

kodxana:
Thanks, the endpoint isolation and empty-planning fixes look good, and the ROCm run now provides real coverage.

One part of the timeout fix still needs adjusting: `container_startup_failure()` only receives `diagnostics.container_lines`. An OCI startup error recorded in `diagnostics.sys_errors`, with empty container output, still becomes STUCK and then SKIP. I reproduced this with an “OCI runtime create failed” entrypoint error, and the summary exited 0 with `on-skip: warn`.

Could you scan both log streams for startup failures and add a regression test for a system-log-only error?

mariiachekmasova-runpod:
container_startup_failure() only seeing container_lines - fixed. It now scans diagnostics.sys_lines + diagnostics.container_lines. Your reasoning is exactly right: a container that was never created has no stdout of its own, so its only trace is on RunPod's side.

One thing I changed beyond your suggestion. Scanning diagnostics.sys_errors would have closed your repro and left four similar cases open, because that list is already narrowed by SYS_LOG_ERROR_PATTERN, which matches error/fail/crash wording only:

yes  OCI runtime create failed: container_linux.go:380
yes  ... exec user process caused: exec format error
no   exec: "/start.sh": no such file or directory
no   permission denied while trying to run /start.sh entrypoint
no   /start.sh: line 12: python3: command not found
no   Out of memory: Killed process 1234 (python3)
So PodDiagnostics now also carries the unfiltered system log, filled from the same single API request. sys_errors stays as the display subset, and both it and filter_sys_errors() now say in their docstrings that they must not be used to classify.

Three regression tests in StartupFailureVisibleOnlyInTheSystemLog: your OCI case with empty container output; the four markers the display filter drops (asserting both that the filter drops them and that classification still catches them, so the test keeps its meaning if that pattern is ever widened); and the inverse — an ordinary system log showing a pull and hand-off still classifies as STUCK, not FAIL. Verified all three fail with the one-line fix reverted.

Green on [run 34338141863](https://github.com/runpod/containers/actions/runs/34338141863) — 26 checks, including the unit suite at 55 tests.

MadiatorLabs:
Thanks, rechecked the fix. Scanning the unfiltered system log covers the remaining startup-failure case, and all 55 unit tests passed locally. My outstanding concern is resolved. Looks good to merge from my side.

Madiator2011Work (approved):

PR #164, opened 2026-09-11
Feat: CON-1459 cuda 13.2, ComfyUI v0.35

mariiachekmasova-runpod:
Adds CUDA 13.2 across the base, PyTorch, cluster and ComfyUI templates.
`versions.hcl` gains 13.2.0 (Ubuntu 24.04 only), and each family gets a
`cu1320` / `cuda132` target set plus a CI matrix entry. The cluster
template picks up four rows (torch 2.9.1, 2.12.0, 2.12.1, 2.13.0).

## PyTorch template

Eight 13.2 rows. torch 2.6–2.9.1 keep taking wheels from the older
indexes they already use; 2.12.0, 2.12.1 and 2.13.0 use cu132.

torchaudio moves out of the `TORCH` spec list into its own install step,
`--no-deps`, with its own index selected by a new `audio_src`. torchaudio's
last release is 2.11.0 and the cu132 index stops at 2.2.0, so 13.2 takes
2.11.0 from cu130 — same CUDA major, and the wheel declares no
dependencies, so it cannot move this build's torch.

## ComfyUI template

New `cuda13.2` variant, and the only one on the new stack: torch 2.13.0 /
torchvision 0.28.0, because the cu132 index has nothing before 2.12.0.

12.8 and 13.0 stay on torch 2.10.0 on purpose. torch 2.11 dropped Volta
(SM 7.0), 2.13 changed the C++ ABI, and torchvision 0.26 removed the video
I/O — leaving the older variants alone confines all of that to a variant
nobody is running yet.

torchaudio on 13.2 is `2.11.0+cpu`. Every CUDA build of 2.11.0, including
the one on PyPI, is cu130, and `torchaudio._check_cuda_version()` raises
when its CUDA does not match torch's — this took the pod down outright.
The `+cpu` wheel reports no CUDA, so the check is skipped. Nothing is
lost: after 2.9 moved I/O to torchcodec, the remaining transforms are
torch ops that still run on the GPU (verified on a pod — resample and mel
spectrogram on `cuda:0`).

13.2 owns its own venv on the network volume, `.venv-cu132`, via a new
`COMFYUI_VENV_NAME` build arg; 12.8 and 13.0 keep sharing `.venv-cu128`,
so existing volumes are untouched. Moving a volume between variants is
handled in `start.sh`: it detects a venv from another variant, reinstalls
user custom-node dependencies into the new one, leaves the old venv in
place so going back still works, and prints how to list hand-installed
packages and how to delete the old venv to free volume space.

Pins bumped: ComfyUI v0.30.0 -> v0.35.0, plus the Manager and KJNodes SHAs.

## Testing

Volume migration 13.0 -> 13.2 verified on a live pod: the existing install
was recognised (no first-time copy), custom-node dependencies were
reinstalled into `.venv-cu132`, hand-installed packages were not carried
over and were reported as such, `.venv-cu128` was left with a removal
hint, torch `2.13.0+cu132` and torchaudio `2.11.0+cpu` resolved from the
image, and ComfyUI started.


kodxana:
The CUDA 13.2 PyTorch rows still use audio_src = "130". TorchAudio checks the CUDA minor version too, so importing it alongside cu132 Torch raises a version mismatch. ComfyUI already fixes this with 2.11.0+cpu; could we apply a compatible solution to the PyTorch images and verify import torchaudio plus an audio operation? The cluster images inherit this combination too.

PR #168, opened 2026-09-18
fix: pin tornado>=6.5.10 (JupyterLab blank page on 1.3.1) and smoke-test a static asset

rambo-runpod:
## Problem

JupyterLab loads a blank white page on `1.3.1` pods. Every `/static/...`
request returns 500:

```
AttributeError: 'FileFindHandler' object has no attribute 'allowed_symlink_directory'
```

tornado 6.5.9 added `allowed_symlink_directory` to `StaticFileHandler`.
jupyter_server 2.21.0's `FileFindHandler` overrides `initialize()` without
calling `super()`, so the attribute is never set.

Upstream: https://github.com/tornadoweb/tornado/issues/3724

Reported by a customer via PostHog feedback:
https://runpod-team.slack.com/archives/C087VB1SFAA/p1789713148194469

## Why 1.3.1 caught it

tornado 6.5.9 published 2026-09-14 18:08Z, 6.5.10 at 2026-09-15 13:47Z.
The 1.3.1 images built at 13:26-13:30Z, inside that ~19.5h window by
under 20 minutes. 1.3.0 is clean.

Verified by reading dist-info out of the published layers:

| image | tornado | state |
|---|---|---|
| `base:1.3.1-*` (7 non-rocm tags) | 6.5.9 | broken |
| `autoresearch:1.3.1-*` (2) | 6.5.9 | broken |
| `comfyui:1.3.1-*` (2) | 6.5.9 | broken |
| `base:1.3.1-rocm644-*` (3) | 6.5.7 | see below |
| `nvidia-pytorch:1.3.1-25.11` | 6.5.8 | ok |

## Fix

tornado reached these images four different ways, so four pin sites:

- `base/requirements.txt` was unpinned
- `comfyui/Dockerfile` resolves `jupyter` through `pip-compile`
- `nvidia-pytorch/requirements.txt` was at 6.5.8
- `rocm/requirements.txt` was at 6.5.7

## This also closes SEC-130

SEC-130 (HIGH, due 2026-09-23) asks to pin tornado to the release
carrying `de85b3f87446` and rebuild. That commit first ships in 6.5.8,
verified against the tornado tags. The rocm images at 6.5.7 are the
family still exposed to CVE-2026-91990. 6.5.10 clears it everywhere.

SEC-130 lists two items as Unverified. Both are answered above: the
fixed release is 6.5.8, and the shipped versions are in the table.

## Why CI passed

The Jupyter smoke test only probes `/api/status`, which is served by a
plain handler and stays 200 while every static route 500s. Reproduced
locally:

```
jupyter_server 2.21.0 + tornado 6.5.9
  /api/status          -> 200
  /static/favicons/... -> 500 text/html

jupyter_server 2.21.0 + tornado 6.5.10
  /api/status          -> 200
  /static/favicons/... -> 200 image/x-icon
```

So no pin would have caught a future recurrence either. Added
`run_jupyter_static_check`, which fetches a static asset after the status
probe passes and rejects an HTML content-type so a login-page redirect
cannot false-pass. 4 unit tests; full unit suite 67 passed.

## Not in this PR

Tags are version-pinned and cosign-signed per digest, and the release
pipeline derives the version from conventional commits, so there is no
supported way to rebuild 1.3.1 in place. This needs a 1.3.2 release plus
repointing the affected pod templates. 1.3.0 is a clean rollback target
in the meantime.

Separately: `rocm.yml` and `comfyui.yml` call the grype action without
`fail-on-findings`, so only base and nvidia gate on CRITICAL/HIGH. That is
how rocm drifted to a June tornado carrying a HIGH. Worth its own ticket.


mariiachekmasova-runpod (approved):
LGTM. There's one thing I'd add - retries for the static check, but since this PR is urgent, I'd merge it as is and add the retries in a follow-up PR. Approved from my side.

PR #169, opened 2026-09-18
fix: don't mint a per-boot JupyterLab token in the ComfyUI container

justinwlin:
## What's broken

Start a ComfyUI pod. Open JupyterLab. It asks for a token. Restart the pod. It asks for a **different** token. Every single time.

The only place that new token is written down is the pod logs, so users have to go read the logs again after every restart just to open Jupyter.

## Why

JupyterLab needs a token. It can come from two places:

1. **The platform.** When you deploy with "Start Jupyter notebook" checked, Runpod makes a token, saves it on the pod, and never changes it. The console knows this token, so the Connect button just works and you never see it.
2. **Our startup script**, if the platform didn't set one.

Today the ComfyUI script does #2 with `openssl rand -hex 16` — a brand new token on every boot. That's the rotation.

Before #163 the script started Jupyter with an empty token, which means **no password at all** on a public URL. #163 was right to close that. It just made the replacement token throwaway, copying the `setup_ssh` root-password fallback a few lines above. Fine for SSH, where nothing else needs to know the value. Wrong for Jupyter, where the token *is* the daily login.

## The fix

Follow what the pytorch and base templates already do (`container-template/start.sh:77`): **no token, no Jupyter.** Don't invent one.

A token minted inside the container can never reach the pod env, so the console can't use it — it's a shell variable in one function and then it's gone. Generating one only creates the rotation.

## Plus: an explicit way to turn auth off

Some people genuinely want a password-less notebook, and this pod already serves **ComfyUI itself unauthenticated on `:8188`**, so the security model here is already "the proxy URL is the secret."

`JUPYTER_DISABLE_AUTH=true` starts JupyterLab with no password and logs two warnings.

It's a separate variable, **not** an empty `JUPYTER_PASSWORD`, on purpose: an empty `JUPYTER_PASSWORD` is exactly what shipped before #163 and left Jupyter open by accident, so that value has to keep meaning "don't start." Compared against the literal string `true`, so a stray `1` or `yes` can't trip it.

## Behavior

| Env | Result |
|---|---|
| `JUPYTER_PASSWORD` set | Jupyter with that token — console Connect button works |
| nothing set | Jupyter doesn't start, logs say why |
| `JUPYTER_DISABLE_AUTH=true` | Jupyter, no password, two warnings in the logs |

**Default console deploys are unaffected.** Both official ComfyUI templates (`cw3nka7d08`, `wgd3p4n4o6`) ship `startJupyter: true`, so the platform always sets `JUPYTER_PASSWORD`, Jupyter always starts, and Connect is one click.

## Who does change

Someone who creates a pod from the **raw image** without `startJupyter`, without a `templateId`, and without setting `JUPYTER_PASSWORD` themselves — or who runs the image with plain `docker run`. They get no Jupyter instead of a log-only rotating token.

That's already what a `runpod/pytorch` pod does for that same user today, so this makes ComfyUI consistent rather than special. Worth calling out in review, since it is a change from what ships right now.

Everything else is covered:

- **Console** — templates ship `startJupyter: true`, backend injects `JUPYTER_PASSWORD`.
- **v2 REST / MCP / runpodctl with `templateId`** — `rphttp2` seeds `StartJupyter` from the template (`internal/templates/expand.go:202`) and forwards it (`internal/pods/translate.go:450`), so the platform injects the variable the same way.
- **v2 REST with an explicit `startJupyter: true`** — same path.

<details>
<summary>How our other templates handle a missing secret</summary>

Four patterns already in this repo:

| Pattern | Where | No secret → |
|---|---|---|
| **Don't start** | `container-template/start.sh:77` — pytorch, base, rocm, autoresearch | Service never starts |
| **Fail closed** | `pytorch-cluster` Grafana proxy, `auth_proxy.py:59` | Service runs, every request denied |
| **Generate per boot** | `comfyui` `setup_ssh`, root password | New value each boot, logged |
| **Initialize once, persist** | `comfyui` FileBrowser, `start.sh:316` | Set on first boot, stored on the volume, stable after |

#163 applied the third to Jupyter. This PR moves it to the first, which is where every other official template already is.

</details>

<details>
<summary>What a pod with no <code>JUPYTER_PASSWORD</code> did, over time</summary>

| | Result |
|---|---|
| ComfyUI before #163 | Jupyter starts, **wide open** |
| ComfyUI today | Jupyter starts, **new token every boot**, only in the logs |
| ComfyUI with this PR | Jupyter doesn't start, logs say why |
| ComfyUI with this PR + `JUPYTER_DISABLE_AUTH=true` | Jupyter starts wide open, **on purpose**, warned |

</details>

## For users hitting this now

Set a `JUPYTER_PASSWORD` env var on the pod to any value. It never changes, and the console Connect button starts working.

### Live pod

Image from this branch pushed and deployed on a real Runpod pod (A40, CA-MTL-1, terminated after the run). Created through the **v2 REST API with `startJupyter: true`** and no explicit password, which is the same injection path the console uses.

The platform put `JUPYTER_PASSWORD` in the pod env — 20 chars of `[0-9a-z]`, the backend's `generateRandomPassword` format, not the container's 32-hex `openssl rand`. Probing the public `:8888` proxy:

| Request | Result |
|---|---|
| `/api/status` no token | **403** |
| `/api/status?token=wrongwrongwrong` | **403** |
| `/api/status?token=<platform token>` | **200** |
| `/lab?token=<platform token>` (the console Connect URL) | **200** |

Only the platform-injected value authenticates, so that is necessarily the token Jupyter launched with — no container-generated token involved. `/api/contents` confirms the root is `/workspace`. ComfyUI came up on 8188 in the same boot.

### In-container

Image built for `linux/amd64` from this branch (Depot, unmodified Dockerfile) and all cases run **inside the real container** against the shipped `/start.sh`:

| Env | Jupyter | `/api/status` no token | with token |
|---|---|---|---|
| `JUPYTER_PASSWORD=realpw123` | running | **403** | **200** |
| nothing set | **not running** | — | — |
| `JUPYTER_DISABLE_AUTH=true` | running | **200** | 200 |
| `JUPYTER_DISABLE_AUTH=yes` | **not running** | — | — |

- 403 → 200 confirms auth is actually enforced in the normal case, not just a token string on the command line.
- 200 with no token confirms `JUPYTER_DISABLE_AUTH=true` genuinely disables auth in the JupyterLab version this image ships.
- `yes` does not disable auth, confirming the strict `= "true"` match.

The bug itself reproduces in the same image using `main`'s script — three boots, three tokens:

```
boot 1 token: cee47044d45e994753b89994328e96ce
boot 2 token: 09485b9d9a0b3e43116477e368769d6f
boot 3 token: 4eb6d44c0cf00f31416ce84c861aca8d
```

Same 32-hex-char shape the reporting user saw in their Jupyter process.

## Checks

- `shellcheck -S warning` clean — the level `.github/workflows/shellcheck.yml` fails at.
- Full image build: 0 errors, all stages `DONE`, `COPY scripts/start.sh /start.sh` included.
- The early return is `return 0`, not `return 1`. The script runs under `set -e` and calls `start_jupyter` at top level, so a non-zero return would abort the pod boot.
- CI still exercises a real Jupyter server: `.github/workflows/comfyui.yml:152` sets `test-jupyter: "true"`, which makes the harness inject `JUPYTER_PASSWORD` (`tests/runpod_smoke/pod.py:220`), so smoke tests take the env path.
- README port table updated.
- Live pod run completed (see above); the pod has been terminated.

## Follow-ups, not in this PR

- `runpod-plugins-official` `runpod-templates/reference/comfyui.md:68` should document the gate once a new image ships. The sibling `pytorch.md:46` already words it correctly.


mariiachekmasova-runpod (commented):
Overall the fix looks good. Three things I'd change:

1. If the goal is for ComfyUI not to diverge from the other templates, then JUPYTER_DISABLE_AUTH should be supported there too.
2. The README should mention that JUPYTER_DISABLE_AUTH takes precedence over JUPYTER_PASSWORD, in case someone sets both JUPYTER_DISABLE_AUTH=true and JUPYTER_PASSWORD at once.
3. Minor: I'd shorten the comment above the if to 2–3 sentences.

container-template/start.sh
```diff
@@ -73,21 +73,38 @@ export_env_vars() {
...
-            fi
+    local JUPYTER_TOKEN
+    if [[ ${JUPYTER_DISABLE_AUTH:-} == "true" ]]; then
+        echo "WARNING: JUPYTER_DISABLE_AUTH=true -- starting JupyterLab with NO authentication."
+        echo "WARNING: anyone with this pod's :8888 proxy URL gets a root shell and full access to /workspace."
+        JUPYTER_TOKEN=""
+    elif [[ $JUPYTER_PASSWORD ]]; then
+        JUPYTER_TOKEN="$JUPYTER_PASSWORD"
+    else
+        return 0
```
mariiachekmasova-runpod: Could you add a note that JUPYTER_PASSWORD isn't set ... , like you did for ComfyUI?

